PR #5, opened 2020-11-12
Update Tests to Core3.1

AlanGraham:
Core3.1 has LTS.
-you could consider changing to .Net5.0 but will need changes to your CI/CD.
Update nuget packages.
Ran Test coverage and dotnet-stryker (https://github.com/stryker-mutator/stryker-net).
-quite low test coverage
-mutation testing with stryker showed a number of areas tests don't cover
-format of tests was not one I'm familiar with and mostly in German so bit of a barrier to submit for now

Cool project - was wrestling with a problem last night and realised Fractions may be the fix.
Had a look and saw this existed, so will try and use to see if it helps :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-bin/
+﻿bin/
 obj/
 _build.output/
 _build.output/*
@@ -17,3 +17,4 @@ _Resharper*
 build.log
 Fractions.XML
 Fractions.userprefs
+/TestResults/*
--- a/src/Fractions.Json/Fractions.Json.csproj
+++ b/src/Fractions.Json/Fractions.Json.csproj
@@ -49,9 +49,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0-preview.2" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
     <PackageReference Include="SourceLink.Create.CommandLine" Version="2.8.3" PrivateAssets="All" />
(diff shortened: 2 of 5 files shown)


AlanGraham:
Had meant to merge to the Main on my fork first, can't see where in files in fork to change the CI to use 3.1 instead of 2.2

danm-de:
Cool, thanks. I'll have a look at the CI/DI pipeline.

danm-de:
I've just checked the code-coverage:

image

I don't know how to convince Resharper's unit test coverage tool to run the tests for all frameworks (`Fractions` is a multi-target project). However, 71% is not that bad ;-)

danm-de:
Unit-Test format is basically AAA (Arrange, Act, Assert). Sometimes a `Spec` base class is used. The virtual methods
- Arrange() or SetUp() 
- Act()

are self-describing (I think). The public `[Test]` method is the assertion step. NUnit's `TestCaseSource` feature is used as well to simplify / minimize code.

I tried to name the classes and methods to read the test results like this:

```
   If something happens ... <class name>
     When it should ... <test method name>
```

Sorry for the Germany language in some of the unit tests ;-)

AlanGraham:
Ah yeah I'd used the inbuilt VS code coverage tool which gives similar results (https://docs.microsoft.com/en-gb/visualstudio/test/using-code-coverage-to-determine-how-much-code-is-being-tested?view=vs-2019 but looks like it's only in Enterprise :( )

CodeCoverage

The mutation report does a better job of showing the gaps in testing.

MutationReport
image

I'm still working on the issue that made me look into Fraction, so once I get as far as I can go, I'll come back to this and maybe spend some time looking more at the tests. I think the first once I'd tried to look at was overriding methods in your Spec class, using Test Sources and not in a language I speak so I didn't want to spend too long trying to understand it before I continued on the tasks I am on. I'm probably more used to https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-nunit style

PR #24, opened 2024-03-30
Added the CLSCompliant(true) attribute

lipchev:
I'm trying to reference this great library in a CLS Compliant [project](https://github.com/angularsen/UnitsNet/pull/1377) and getting blasted with CS3003 warnings, so I thought it wouldn't hurt to have it added here instead.


lipchev:
It seems like the AppVeyor is struggling with some weird characters in the test names (I don't mean the accents :)), but hopefully it should be fine after/if #25 is merged.

danm-de (approved):
Thanks!

PR #30, opened 2024-04-10
Make the operator from decimal implicit

lipchev:
Was there a reason for this not being an implicit operator?

PS.
Should also update the gh-pages branch.


danm-de (approved):
Since we don't lose any precision in the conversion, I can live with 'implicit'.

PR #31, opened 2024-04-10
Add the Fraction.Round functions (prototype from decimal.Round)

lipchev:
I've added to Round(..) functions to the Fraction type itself (same as the decimal), but if you prefer we could move it to the MathExt class.. 

Note that the `MidpointRounding` enum [was extended](https://learn.microsoft.com/en-us/dotnet/api/system.midpointrounding?view=netcore-3.0) in .NET Core 3.0.
I have added the corresponding `#if NET` check and throwing an exception if the mode is not available, but this only makes sense if we actually bump the project version (otherwise the associated tests would fail). 

I have therefore changed the netstandard2.1 target to net7.0 (the app-veyor image needs to be update in order to support net8.0).

### DIFF
--- a/src/Fractions/Fraction.Math.cs
+++ b/src/Fractions/Fraction.Math.cs
@@ -192,4 +192,95 @@ public readonly partial struct Fraction {
                 numerator: fraction.Denominator,
                 denominator: fraction.Numerator,
                 state: fraction.State);
+
+    /// <summary>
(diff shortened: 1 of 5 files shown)


danm-de:
Thank you!

I'm going to make a few changes:
- Changing the framework to .NET 8.0,
- The rounding method that returns BigInteger will be renamed to `RoundToBigInteger`.

PR #32, opened 2024-04-11
Extended the Pow/Sqrt functions with their fractional-exponent overloads

lipchev:
Similar to `Math.Pow(double x, double y)` but with the added accuracy parameter..

I'm really not sure about this one...when used "unwisely" - there are many ways this could lead to an overflow / never-ending loop..

Personally, I don't think I will ever need this, and was in fact about to delete it- but if you want it (or if it ever comes up again later), well here it is..

### DIFF
--- a/src/Fractions/Extensions/FractionExt.cs
+++ b/src/Fractions/Extensions/FractionExt.cs
@@ -44,4 +44,49 @@ public static class FractionExt {
 
         return newGuess;
     }
+
+    /// <summary>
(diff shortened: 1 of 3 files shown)


danm-de:
You already said it yourself [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

I'm going to close this pull request for now - if anyone still needs it, we can think about it again.

PR #35, opened 2024-04-11
Lipchev string parsing with scientific notation

danm-de:


lipchev:
I'll probably have a run of the benchmarks later, but my guess is that the `ReadOnlySpan<char>` version is going to perform significantly better (without factoring the performance improvements in the framework itself). 

If there is indeed a meaningful gain (or you simply would want to reduce the code duplication), there is always the possibility of referencing the [System.Memory nuget](https://www.nuget.org/packages/System.Memory). It is another 204.08 KB (haven't checked the actual dll, just reading off nuget.org), but I have the feeling that just about every library out there is now referencing it anyway...

In any case, this isn't something that needs to be decided right now, just wanted to bring it up in case you weren't aware of it (I personally never realized what the nuget was for up until a few months ago...)

lipchev:
Oh, sorry - I hadn't realized how small the Fraction package is in comparison.. :)

PR #36, opened 2024-04-12
Implement DecimalNotationFormatter and associated tests

lipchev:
Description:
- Added a new `DecimalNotationFormatter` class for formatting Fraction objects into decimal string representations corresponding to the  [Standard numeric format strings.](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings)
- Created tests for the `DecimalNotationFormatter` in the `DecimalNotationFormatter_Format` 

Here are the supported format options (with _unlimited_ precision):
- [x] The general ("G") format specifier converts a number to a string that is formatted as the most compact of either fixed-point or scientific notation.
- [x] The scientific ("E") format specifier converts a number to a string of the form "-d.ddd…E+ddd" or "-d.ddd…e+ddd", where each "d" indicates a digit (0-9). The string starts with a minus sign if the number is negative.
- [x] The fixed-point ("F") format specifier converts a number to a string of the form "-ddd.ddd…" where each "d" indicates a digit (0-9). The string starts with a minus sign if the number is negative.
- [x] The standard numeric ("N") format specifier converts a number to a string of the form "-d,ddd,ddd.ddd…", where "-" indicates a negative number symbol, "," indicates a group separator, and "d" indicates a digit (0-9). The string starts with a minus sign if the number is negative *.
- [x] The percent ("P") format specifier multiplies the number by 100 and converts it to a string that represents a percentage.
- [x] The currency ("C") format specifier converts a number to a string that represents a currency amount.
- [x] The "R" format specifier attempts to ensure that a numeric value that is converted to a string is parsed back into the same numeric value: _implemented via casting to `double` (maintaining interoperability at the cost of precision)_
- [x] The significant digits after radix ("S") format specifier converts a number to a string that preserves the precision of the Fraction object with a variable number of digits after the radix point. 
- [x] Custom format: _these are only supported using the lossy ToDouble() conversion_
- [x] Update the README.md to include another section about the `DecimalNotationFormatter` 
- [x] Add benchmarks: _the results appear to be within a magnitude difference of the equivalent `ToDouble().ToString(...)`_


danm-de:
That's pretty cool. I wish this article from Microsoft had existed years ago (or it hadn't been so well hidden). Then I would have used it as a guide for the `ToString` method.

I find this extension very useful. Unit tests are always valuable and necessary. However, this class would at least have to be mentioned and documented in the README.md. 😄

lipchev:
Just a couple of points to clear up:

1. What do you think of the name, isn't it a bit ambiguous - I mean we are "using the decimal format to format a fraction" not necessarily "formatting a decimal fraction".. Any suggestions are welcome...
2. What do you think about the "S" format- it is the "standard" format we've been using in UnitsNet. It isn't based on any _official_ format as far as I'm aware.
3. What do you think about the "R" format, although we can easily support the "a/b" formatting (like the DefaultFractionFormatter) I think we should keep it as "GXX" - but what value should we select for the precision ("XX")?

danm-de:
1. Ohha, the good old [naming problem](https://www.amazon.de/Naming-Things-Hardest-Software-Engineering/dp/B0BTLYZWRL). I will try it:
- `DecimalNotationFormatter` (fully qualified with namespace Fractions.Formatter**s**.DecimalNotationFormatter)
- `DecimalFormatter` (fully qualified with namespace Fractions.Formatter**s**.DecimalFormatter)
I know, pretty unimaginative and boring.

2. S-Format: I'm not entirely sure - but I think my colleagues chose similar formatting for the WPF UI of the medical software. For dosage information, the essential digit after the decimal point must always be displayed. I rely on your expertise - if you think it's good and necessary, feel free to add it. The letter is not yet used in the standard defined by Microsoft. So I don't see any problems with that.

3. As I understand the [documentation](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#RFormatString), the round-trip format specifier is intended to guarantee that the string representation can be parsed back to exactly the same value. I therefore assume that this specifier is only used for purely technical applications (e.g. serialization). Although this formatter is written specifically for decimal - how do you want to ensure that 1/3 (0.33333...) can be parsed back to the exact `Fraction` value? I think that only works in fractional notation. The documentation also states that the precision specifier is ignored. Personally, I would choose <numerator>/<denumerator> (e.g. `"1/3"`). An interesting fun fact: the specifier for the `DateTime` data type is actually a small `o` instead of `R`. You actually always get the correct date - even with the time zone.
image

As I wrote before - I find this formatter pretty cool and useful. My current implementation can only output decimal numbers in the representable range of the `decimal` data type. However, without instructions or mention in the documentation (README.md?) it will not be usable for most people (because they don't know it exists).

danm-de:
If I haven't said it yet: thank you for your commitment and dedication. I really appreciate that.

tests/Fractions.Tests/StringFormatters/DecimalNotationFormatter_Format.cs
```diff
@@ -0,0 +1,822 @@
...
+        // Arrange
+        var doubleToString = valueToTest.ToDouble().ToString(stringFormat, formatProvider);
+
+        // Act
+        var fractionToString = DecimalFormatter.Format(stringFormat, valueToTest, formatProvider);
+
+        // Assert:
+        // while the fraction is correctly rounded to MidpointRounding.ToEven,
+        // there is apparently some optimization done to the double.ToString("g") that give weird results on the last decimal (cannot find any reference)
+        // (7.0/3).ToString(CultureInfo.InvariantCulture) returns "2,3333333333333335"
```
danm-de, line 104: This is due to the representation of 7/3 when using floating point arithmetic according to IEEE 754. Since we do not store the number in base 10, it is stored in an (unfortunately finite sequence, (1 + 2^(-1) + 2^( -2) .. + 2^(-52)) * 2^(exponent-K))

The rounding method inside `double` works correctly:

image

lipchev: The puzzling thing here was not related to our code, but rather the behavior of ToString with non-terminating decimals like (7.0/3) between .NET 48 and .NET8 - in the first case the value is always printed with the correct rounding, where as .NET8 appears to have "optimized" something and i cannot find when or what exactly (when using a precision specifier both frameworks act the same way) 😄 

> That's pretty cool. I wish this article from Microsoft had existed years ago (or it hadn't been so well hidden)

PR #41, opened 2024-04-16
Introduced the negation (-) operator

lipchev:
Added the negation operator (implemented using the Invert operator)

<!DOCTYPE html>
<HTML>
<body>

  | Name | Value | Type
-- | -- | -- | --
  | Fraction.MinusOne == -Fraction.One | true | bool

</body>
</HTML>


lipchev:
By the way, I believe that term "inverting" is typically used for "finding the reciprocal of a fraction". What we're doing in the "Invert" method is "negating" the fraction.
If you agree, I think that it would be best if we made the method [Obsolete] with a warning message such as 

> The 'invert' method is obsolete. Use the negation operator '-x' or the 'Negate' method to multiply the fraction by -1.

That is if we want to have the method overload

Issue #26, opened 2024-03-30
Support for NaN and Infinity


lipchev:
## Support for NaN and Infinity
Hey,
I've been digging into the code and I think we can add support for NaN and +/- Infinity. Here's what I'm thinking:
1. Update `GetReducedFraction(..)` to return 0/0 (NaN) for 0 in the numerator and denominator. Right now, it's returning `Zero`.
2. For a positive numerator and 0 in the denominator, return +1/0 (+Inf).
3. For a negative numerator and 0 in the denominator, return -1/0 (-Inf).

We might need to tweak `FractionState` to `{Zero, Unknown, Normalized}` to avoid `default(Fraction)` being NaN. Also, in that case, `Reduce()` should check for `FractionState.Zero` and return a _true_ `Zero` (0/1).
There might be a few more checks needed (like in the constructor from `double`), but I think it's doable. 

I'm ready to make these changes and submit a PR if you're on board with this.

Let me know what you think!

danm-de:
Thanks for this suggestion. Perhaps I should say something about the history of this project. Fractions started as a private PoC for calculating drug dosage for clinical applications. The software of my employer at the time had major problems in displaying calculated volume rates, active ingredient rates and summations of critical active ingredient quantities over a specific period of time.

(Sorry for the German language in the pictures - I just paste parts of my old slides ;-)

![image](https://github.com/danm-de/Fractions/assets/3825484/a751a8ab-987e-4d56-8b30-c359c6b943e3)

The data type _double_ has too large rounding errors and is therefore disqualified for drug dosages prescribed in the nanogram range. The data type _decimal_ is less susceptible to the typical errors of floating-point arithmetic due to the base of 10 - BUT - periodic decimals (which could occur briefly during computation) continued to be a problem.

![image](https://github.com/danm-de/Fractions/assets/3825484/ebaf88c6-2f0d-48d5-8fb3-6b5f8685c6b3)

So, the rounding errors were still an issue when using the _decimal_ data type. And the doctors and nurses freaked out... Instead of an infusion rate of 5mg/kg/h, the UI said 4.997mg/kg/h OR even worse 0.998 pill instead of simply 1 pill.

![image](https://github.com/danm-de/Fractions/assets/3825484/4f127289-97a1-46e8-bca0-356d50203020)

[...]

PR #48, opened 2024-04-21
Adding support for NaN and Infinity

lipchev:
As discussed in #26 
- [x] NaN represented by {0/0}
- [x] PositiveInfinity represented by {1/0}
- [x] NegativeInfinity represented by {-1/0}
- [x] Zero is now represented by {0/1}
- [x] default(Fraction) results in {0/1} (with `FractionState.Unknown`)
- [x] Supporting all types of non-normalized fractions such as: {10/0} (+Infinity), {0,-10} (Zero), {-10/0} (-Infinity)
- [x] Benchmark results show an all-around performance improvement * (_note these aren't so much related to the introduction of NaN, but more as a result of the more granular handling of the edge-cases_)

I ended up not changing the `FractionState` enum, instead it is being check for when returning the _default_ `Denominator` 

Breaking changes:
- `Fraction.Zero.Denominator` now returns `1`
- no exceptions during construction
- since it is now possible to construct fractions with 0 in the `Denominator` - functions such as `ToInt64()` or `ToDecimal()` would throw a `DivideByZeroException`
- according to the same logic `Fraction.NaN.ToString("n")` would also throw a `DivideByZeroException`
- _I'm not sure if before it was possible to parse `"0/-10"` as a non-normalized `Zero`, and then have it `ToString()` back to it's original value `"0/-10"`_


lipchev:
There's a lot to review, I'm sorry- I didn't mean to change _every single file_ but there were a lot of case to check for especially when it comes to the non-normalized fractions. I initially didn't consider having them supported for the positive/negative infinity, but as i reached the `string <-> Fraction <-> string` scenarios I thought, whatever- let's have all possible "a/b" cases covered..

I've left (myself) a few `TODO`s with possible shortcuts to take, but wanted to have _something_ to test them against (the benchmarks that I uploaded with this release). 

While updating/testing the code, I also noticed that there isn't a consistent rule regarding the `FractionState` of operations such as `Multiply`/`Divide`/`Reciprocal` - `most` of the time it appears we're forcing a normalization, but not always..

src/Fractions/Fraction.Operators.cs
lipchev, line 1: There is a special rule regarding the `Equals` method and the `==` operator w.r.t `double.NaN`:
https://learn.microsoft.com/en-us/dotnet/api/system.double.nan?view=net-8.0#remarks

src/Fractions/Fraction.cs
lipchev, line 1: I don't really like how `return _nan;` reads, do you think it's worth having the backing fields?
I had to switch to using the `Numerator` and `Denominator` properties, but this was so that I can have the _default-value-handling-logic_ executed in the getter (and felt like mixing `_numerator` and `Denominator` in the same call isn't worth it..

An alternative implementation (which might have been slightly more readable) is to have the `_denominator` defined as `Nullable<BigInteger>` and returning `_denominator.GetValueOrDefault(BigInteger.One)` in the getter.

danm-de:
Thanks for the pull-request.

Unfortunately, due to time constraints, I need a little longer for the code review this time (a few days).


I still have a quick question:

Is the following condition always met?

```csharp
Fraction.Zero == default(Fraction) 
```

lipchev:
> Thanks for the pull-request.
> 
> Unfortunately, due to time constraints, I need a little longer for the code review this time (a few days).
> 
> I still have a quick question:
> 
> Is the following condition always met?
> 
> ```cs
> Fraction.Zero == default(Fraction) 
> ```

Yes, that should evaluate as `true`

lipchev:
> I've left (myself) a few `TODO`s with possible shortcuts to take, but wanted to have _something_ to test them against (the benchmarks that I uploaded with this release).

I went over these in my last commit, and added (_after substantial amount of benchmarking_) the required edge-case-testing in order to minimize the number of BigInteger allocations (most often due to large multiplications). 
I've specifically tested the effect on using the GCD to reduce the size of the terms added/multiplied- this was only hurting the performance, up to a certain size, after which having two very-large numbers multiplied is significantly more expensive than figuring out their GDC and doing all the extra divisions etc. There is probably some formula that can be determined here, but I didn't go as far. I just tested one against the other for the different cases and went with that:
- CompareTo: since there are no further allocations (no additions/multiplications) using the GCD was always slower (i've removed it)
- Add: here the results were mixed- mostly better without the pre-normalization, except for the very large cases - where the difference was quite significant (i've left it in)
- Remainder: same as with the addition, only here there are even less cases where the GCD is worth it (i've left it in)
- 

It takes a lot more _ifs_ and _elses_, but even operations such as Multiply/Divide/Pow can benefit from not doing two multiplications (when one is expected to return 0):

>         // we want to skip the multiplications if we know the result is going to be 0/b or a/0

Note that these improvements could be applied as "non-breaking-optimizations", if we're committed on having a final/stable release of the v7... We're talking more than 20%-30% improvements on most benchmarks..

src/Fractions/Fraction.ConvertFrom.cs
```diff
@@ -481,12 +481,19 @@ namespace Fractions;
...
-        if (Math.Abs(value - 0.0) < double.Epsilon) {
+        if (double.IsNegativeInfinity(value)) {
+            return _negativeInfinity;
+        }
+
+        if (double.IsNaN(value)) {
+            return _nan;
+        }
+
+        if (Math.Abs(value - 0) < double.Epsilon) { // TODO does this make any sense or can we replace it with a value == 0 check?
```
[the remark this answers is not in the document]
lipchev: Have a look at the comments under the answer that suggest comparing with `Double.Epsiolon`- those were my thoughts exactly:

>  @sonatique Math.Abs(something) < Double.Epsilon evaluates to true only when something is exactly zero, in which case something == 0 would also be true, so why not just use that?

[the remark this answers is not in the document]
danm-de: Years ago Resharper suggested this replacement to me. I can't find it anymore. This current proposal uses `tolerance`, which is obviously larger than "double.Epsilon".

https://www.jetbrains.com/help/resharper/CompareOfFloatsByEqualityOperator.html

danm-de:
Since the changes are very extensive, I have integrated your GIT branch under https://github.com/danm-de/Fractions/tree/lipchev-support-nan-and-infnity

I will apply my changes directly there, rather than using GITHUB's comment function. Reason: many changes have an impact on many (sub)functions and unit tests.

I'm currently trying your suggestion regarding the `Nullable<BigInteger>` for the denominator. The special handling for the `default(Fraction)` case is too complicated for me or I find the safety tests too difficult to understand 😄

src/Fractions/Fraction.Operators.cs
danm-de: Interesting. I guess it's okay to behave exactly like the double data type.

src/Fractions/Fraction.cs
danm-de: Yes, I'll try this in https://github.com/danm-de/Fractions/tree/lipchev-support-nan-and-infnity

danm-de (commented):
I've finished the code review and have published my suggested changes directly to the following branch:

https://github.com/danm-de/Fractions/tree/lipchev-support-nan-and-infnity

It would be nice if you could take another look.

The main changes are:
- Auto-Properties for the win (backing fields removed if possible)
- Denominator is now nullable
- Correction of the ToString(..) method

src/Fractions/Fraction.ConvertFrom.cs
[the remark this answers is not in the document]
danm-de: Removed, thanks for pointing that out.

danm-de:
> I've finished the code review and have published my suggested changes directly to the following branch:
> 
> https://github.com/danm-de/Fractions/tree/lipchev-support-nan-and-infnity
> 
> It would be nice if you could take another look.
> 
> The main changes are:
> 
>     * Auto-Properties for the win (backing fields removed if possible)
> 
>     * Denominator is now nullable
> 
>     * Correction of the ToString(..) method

Oh, I may introduced a problem with TryParse.

PR #50, opened 2024-04-28
Integration: Lipchev support nan and infnity

danm-de:
Trying to integrate pull request #48 with changes.


lipchev:
I've just pushed a PR (#51) with some minor changes. (replaced one TODO with another 😄 )

src/Fractions/Formatter/DefaultFractionFormatter.cs
```diff
@@ -7,102 +7,130 @@
...
-                "/",
-                fraction.Denominator.ToString(CultureInfo.InvariantCulture));
+            if (fraction.IsNegativeInfinity) {
+                return numberFormatInfo.NegativeInfinitySymbol;
+            }
+
+            if (fraction.Denominator == BigInteger.One) {
+                return fraction.Numerator.ToString(formatProvider);
+            }
         }
```
lipchev, line 129: Note that for `InvariantCultureInfo` the symbols for these are actually longer than the "a/b"  (e.g. "PositiveInfinity").

danm-de:
> I've just pushed a PR (#51) with some minor changes. (replaced one TODO with another 😄 )

Merged.

src/Fractions/Formatter/DefaultFractionFormatter.cs
danm-de: I think as a user I would rather see NaN than 0/0 (or ∞ for infinity) when calling `ToString()`. I introduced another breaking change - the default FormatProvider is now `CultureInfo.CurrentCulture`.

image

lipchev: Sure, that's what I would expect from the `DecimalNotationFormatter`. Personally, I don't mind it either way, I don't think there is any real reason to ever be parsing these anyway 😄

danm-de: Definitely a special case - but `double` has this feature.
image

PR #51, opened 2024-04-29
Nullable denominator review

lipchev:
Ok, I'm now point at the right branch:
- the two additions are fairly straightforward

What do you think about the making `default(Fraction).State == FractionState.IsNormalized`?


src/Fractions/Fraction.cs
lipchev, line 1: This is quicker:
| Method                    | Mean      | Error     | StdDev    |
|-------------------------- |----------:|----------:|----------:|
| NullCoalescing_Null       | 1.1392 ns | 0.0150 ns | 0.0140 ns |
| GetValueOrDefault_Null    | 0.9670 ns | 0.0088 ns | 0.0082 ns |
| NullCoalescing_NotNull    | 1.1386 ns | 0.0124 ns | 0.0116 ns |
| GetValueOrDefault_NotNull | 0.9664 ns | 0.0050 ns | 0.0042 ns |

danm-de (approved):

lipchev:
Oh, no need- I think we can

- switch the order of the the enum states (making IsNormalized the first one)
- keep the enum as it is, but have a boolean backing field (defaulting to `true`) that we use to return the desired `FractionState` in the property getter

danm-de:
> Oh, no need- I think we can
> 
> * switch the order of the the enum states (making IsNormalized the first one)
> * keep the enum as it is, but have a boolean backing field (defaulting to `true`) that we use to return the desired `FractionState` in the property getter

Is there another reason why the _FractionState_ needs to be queried? Would an `IsImproper` property also suffice? 

If so, I would prefer your suggestion with a BackingField (`bool _isImproper` “?). The `State` property remains only for compatibility reasons and is marked obsolete. What do you think?

danm-de:
Unfortunately, the constructor parameter probably needs to be renamed.

```csharp
var a = new Fraction(1, 3, reduce: false);
var b = new Fraction(1, 3, reduce: true);
```

As backing-field: ```_unconfirmedNormalization```?
`default(Fraction)` results in `_unconfirmedNormalization = false`

lipchev:
> Unfortunately, the constructor parameter probably needs to be renamed.
> 
> ```cs
> var a = new Fraction(1, 3, reduce: false);
> var b = new Fraction(1, 3, reduce: true);
> ```
> 
> As backing-field: `_unconfirmedNormalization`? `default(Fraction)` results in `_unconfirmedNormalization = false`
> 
> We leave out the `IsNormalized` property entirely - the benefit is small (if any).

__unconfirmed__ suggests to me that we're awaiting confirmation.. ("where is the `Confirm(..)` method?")
Here's my order of preference:

1. _normalizationDeferred
2. _normalizationPending
3. _normalizationNotApplied

danm-de:
`_normalizationNotApplied` sounds good to me. 

The name of the constructor parameter can be either `normalize` or `reduce`. Come to think of it - "normalize" is an explicit instruction - so not a bad name so far.

lipchev:
> `_normalizationNotApplied` sounds good to me.
> 
> The name of the constructor parameter can be either `normalize` or `reduce`. Come to think of it - "normalize" is an explicit instruction - so not a bad name so far.

As a verb, `normalize` is better, although (if starting clean) I'd personally make anything that's not an implicit constructor, or a variation of the standard tuple (BigInteger, BigInteger) constructor into a factory method. _Especially the constructor from `double`!_

danm-de:
I agree with you. Not all of the design decisions back then were good. However, I would not like to remove these constructors in order to keep the breaking changes within an acceptable range.

lipchev:
Sure, I was thinking of adding some additional comments to all public members but didn't want to pollute the diffs before (we should have done the solution-reformatting stuff earlier...).
Should I make another PR or are you working on it?

danm-de:
The only reason I duplicated your branch in a second pull request was because I lost track of all the changes 🥲 😄 . 

I am currently in the clinic (my employer) and am busy with other topics today.

PR #58, opened 2024-05-04
FromString supporting all NumberStyles

### LINKED ISSUE #56: List of possible issues with string parsing in v7
Here are the summary of the tests that are failing in the current code base:

1. Parsing_a_fraction_from_a_long_decimal_string_with_specific_number_style (23 tests) 
- Should_not_work_with_groups_in_the_middle("123456789987654321.,123456789987654321")
- Should_not_work_with_groups_in_the_middle("123456789987654321.123456789,987654321")
- Should_return_the_expected_value_when_the_string_is_valid("1234,56789987654321.-",Any,)  // all strings with the +/- sign on the right side
- Should_return_the_expected_value_when_the_string_is_valid(" (.1) ",Any,) // all strings with "(..)" as negative pattern
2. Same issues are repeated with the scientific notation
3. All currency formats are affected as well
4. When parsing a string with a culture where the symbols are not the standard length (e.g. having "_minus_" instead of "-" for the negative sign).

When parsing a string with the sign / currency symbol to the right side, the results are incorrect due to the fact that the symbols to the right do not contribute to the number of decimals in the denominator:

```
The_result_should_be_minus_7_over_2("3,5-") Failed: Expected object to be equal to -7/2, but found -7/20.
Expected object to be equal to -7/2, but found -7/20.
   at FluentAssertions.Execution.LateBoundTestFramework.Throw(String message)
   at FluentAssertions.Execution.AssertionScope.FailWith(Func`1 failReasonFunc)
   at FluentAssertions.Numeric.ComparableTypeAssertions`2.Be(T expected, String because, Object[] becauseArgs)
   at Fractions.Tests.FractionSpecs.FromString.When_creating_a_fraction_from_minus_3_5_with_German_culture.The_result_should_be_minus_7_over_2(String value) in C:\Users\galin\source\repos\Fractions-fork\tests\Fractions.Tests\FractionSpecs\FromString\Method_FromString.cs:line 142
```

A PR is coming in a bit..

lipchev:
fixes #56 

Sorry again- way too many weird cases for review...

Benchmark results to come in bit,

Feel free to make changes directly to my branch..

### DIFF
--- a/src/Fractions/Fraction.ConvertFrom.cs
+++ b/src/Fractions/Fraction.ConvertFrom.cs
@@ -879,7 +879,7 @@ public readonly partial struct Fraction {
 
         var nbDecimals = endIndex - decimalSeparatorIndex - decimalSeparatorLength;
         if (nbDecimals == 0) {
-            fraction = new Fraction(false, isNegative ? -numerator : numerator, BigInteger.One);
+            fraction = new Fraction(isNegative ? -numerator : numerator);
             return true;
         }
 
--- a/src/Fractions/Fraction.ConvertFrom.cs
+++ b/src/Fractions/Fraction.ConvertFrom.cs
@@ -51,116 +51,65 @@ public readonly partial struct Fraction {
     /// Try to convert a string to a fraction. Example: "3/4" or "4.5" (the decimal separator character depends on the system's culture).
     /// If the number contains a decimal separator it will be parsed as <see cref="decimal"/>.
     /// </summary>
-    /// <param name="fractionString">A fraction or a (decimal) number. The numerator and denominator must be separated with a '/' (slash) character.</param>
+    /// <param name="value">A fraction or a (decimal) number. The numerator and denominator must be separated with a '/' (slash) character.</param>
     /// <param name="fraction">A normalized <see cref="Fraction"/> if the method returns with <c>true</c>. Otherwise, the value is invalid.</param>
     /// <returns>
-    /// <para><c>true</c> if <paramref name="fractionString"/> was well-formed. The parsing result will be written to <paramref name="fraction"/>. </para>
-    /// <para><c>false</c> if <paramref name="fractionString"/> was invalid.</para></returns>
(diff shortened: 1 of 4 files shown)


src/Fractions/Fraction.ConvertFrom.cs
```diff
@@ -88,26 +89,36 @@ namespace Fractions;
     /// <para><c>true</c> if <paramref name="fractionString"/> was well-formed. The parsing result will be written to <paramref name="fraction"/>. </para>
     /// <para><c>false</c> if <paramref name="fractionString"/> was invalid.</para>
     /// </returns>
-    public static bool TryParse(string fractionString, NumberStyles numberStyles, IFormatProvider formatProvider,
+    public static bool TryParse(ReadOnlySpan<char> fractionString, NumberStyles numberStyles, IFormatProvider formatProvider,
```
lipchev, line 92: Please note that I removed the string version overload in .net8 and renamed the parameter name to match the other one. 
Technically this is a breaking change, but only for people who use reflection (or who have recently switched to using the ReadOnlySpan overload _and_ have named the parameter explicitly).

Personally, I think this is very unlikely to break for anyone, while the advantage (IMO) of not having the `string` overload is that the user would probably notice the little icon indicating an implicit conversion, and consider updating his code with ReadOnlySpan<char>.. 🚀

lipchev:
This is the comparison with `master`. Note that there are 3 cases for which the allocations  (and the mean) in net8 have increased. I have spent a lot of time but could not make them match the base-line. I hope you can spot something that I've missed..

image
image

src/Fractions/Fraction.ConvertFrom.cs
danm-de: I don't think changing the public API really brings much benefit here. [Named Arguments](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments) are used quite frequently (at least by us).

I'll add the overload again for the git-merge.

PR #59, opened 2024-05-05
String parsing hotfix

lipchev:
Same as #58 but targeting the v7 branch.


tests/Fractions.Tests/FractionSpecs/TryParse/Method_TryParse.cs
```diff
@@ -42,11 +42,7 @@ public class When_trying_to_parse_a_fraction : Spec {
     [Test, TestCaseSource(nameof(TestCases))]
     public Fraction? The_result_should_be_as_expected(string value) =>
         Fraction.TryParse(
-#if NET
-            value: value.AsSpan(), // convert to ReadOnlySpan<char>
-#else
-            value, // as string
-#endif
```
lipchev, line 49: Please note this change- I've mentioned it in the other PR as well, but I think this is _slightly_ more problematic for a v7 release (however small the likelihood could be).

danm-de: We must keep the public API surface compatible. SemVer requires us to be binary compatible for minor and patch releases.

PR #60, opened 2024-05-05
Adding a debugger display proxy

lipchev:
Here's what I was thinking:

image

I haven't changed the top-level display (which defaults to the ToString()) implementation, however if we wanted, we could make this some kind of internal `ToShortString()` implementation which uses the decimal notation, when possible, switching to the "a/b" format (or possibly the "m" version) otherwise.

### DIFF
--- a/src/Fractions/Formatter/DecimalNotationFormatter.cs
+++ b/src/Fractions/Formatter/DecimalNotationFormatter.cs
@@ -72,7 +72,7 @@ public class DecimalNotationFormatter : ICustomFormatter {
     /// <remarks>
     ///     This instance can be used to format Fraction objects into decimal string representations.
     /// </remarks>
-    public static ICustomFormatter Instance { get; } = new DecimalNotationFormatter();
+    public static DecimalNotationFormatter Instance { get; } = new();
 
     private static readonly BigInteger Ten = new(10);
 
--- a/src/Fractions/Fraction.DebugProxy.cs
+++ b/src/Fractions/Fraction.DebugProxy.cs
@@ -1,0 +1,31 @@
+﻿using System.Diagnostics;
+using System.Globalization;
+using System.Numerics;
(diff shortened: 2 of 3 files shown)


src/Fractions/Formatter/DecimalNotationFormatter.cs
```diff
@@ -72,7 +72,7 @@ public class DecimalNotationFormatter : ICustomFormatter {
     /// <remarks>
     ///     This instance can be used to format Fraction objects into decimal string representations.
     /// </remarks>
-    public static ICustomFormatter Instance { get; } = new DecimalNotationFormatter();
```
lipchev, line 75: We need this in order for the docs of the `DecimalNotationFormatter.Instance.Format()` method to display our custom formats (not those of the interface).

src/Fractions/Fraction.DebugProxy.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Globalization;
+using System.Numerics;
+using Fractions.Formatter;
+
+namespace Fractions;
+
+public readonly partial struct Fraction {
+    internal readonly struct FractionDebugView(Fraction fraction) {
+        public BigInteger A => fraction.Numerator;
+        public BigInteger B => fraction.Denominator;
```
lipchev, line 11: Was unable to make "Numerator" and "Denominator" display in the order that I wanted (in VS it's alphabetical).

src/Fractions/Formatter/DecimalNotationFormatter.cs
danm-de: Ah, right - sorry.

src/Fractions/Fraction.DebugProxy.cs
danm-de: I like it.

PR #66, opened 2024-05-20
Feature/default to value equality

danm-de:
This pull request introduces a breaking change. The `Equals(..)` method and the `==`, `!=` operators now checks for equality of values.

- [x] New base class `FractionComparer` that implements  `IEqualityComparer<Fraction>`, `IEqualityComparer`
- [x] New class `FractionValueEqualityComparer` (that checks for value equality)
- [x] Data type `Fraction` uses `FractionComparer.ValueEquality` for `Equals(..)` and `GetHashCode()`
- [x] Changed (some) unit tests to use a strict equality check
- [x] New class `FractionStrictEqualityComparer` (that checks the exact values of _numerator_ and _denominator_)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -11,7 +11,8 @@
 
 ### Breaking changes
 
-- Mathematical operations no longer automatically generate normalized fractions if one of the operands is an improper (i.e. non-normalized) fraction. This has an impact on your calculations, especially if you have used the `JsonFractionConverter` with default settings. In such cases, deserialized fractions create improper fractions, which can lead to changed behavior when calling `Equals` and `ToString`. Please refer the README section [Working with non-normalized fractions](Readme.md#working-with-non-normalized-fractions)
+- Mathematical operations no longer automatically generate normalized fractions if one of the operands is an improper (i.e. non-normalized) fraction. This has an impact on your calculations, especially if you have used the `JsonFractionConverter` with default settings. In such cases, deserialized fractions create improper fractions, which can lead to changed behavior when calling `ToString`. Please refer the README section [Working with non-normalized fractions](Readme.md#working-with-non-normalized-fractions)
+- The default behavior of the `.Equals(Fraction)` method has changed: `Equals` now compares the calculated value from the numerator/denominator (1/2 = 2/4).
 - The standard function `ToString()` now depends on the active culture (`CultureInfo.CurrentCulture`). The reason is that NaN and Infinity should be displayed in the system language or the corresponding symbol should be used.
 - Argument name for `Fraction.TryParse(..)` changed from `fractionString` to `value`.
 - A fraction of 0/0 no longer has the value 0, but means NaN (not a number). Any fraction in the form x/0 is no longer a valid number. A denominator with the value 0 corresponds to, depending on the numerator, NaN, PositiveInfinity or NegativeInfinity.
--- a/Fractions.sln
+++ b/Fractions.sln
@@ -43,6 +43,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Results", "Results", "{FCF0
 		benchmarks\results\Fractions.Benchmarks.ExampleScenarioBenchmarks-report-github.md = benchmarks\results\Fractions.Benchmarks.ExampleScenarioBenchmarks-report-github.md
 		benchmarks\results\Fractions.Benchmarks.FromDoubleBenchmarks-report-github.md = benchmarks\results\Fractions.Benchmarks.FromDoubleBenchmarks-report-github.md
 		benchmarks\results\Fractions.Benchmarks.FromStringBenchmarks-report-github.md = benchmarks\results\Fractions.Benchmarks.FromStringBenchmarks-report-github.md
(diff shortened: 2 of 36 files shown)


danm-de:
> @lipchev Found a small bug that was introduced when switching to non-normalized operations. The following special case currently produces incorrect results:

Simple fix 030df28901fd2254e776108cd2be0103da6ad04d

lipchev:
>     * [ ]  New class `FractionStrictEqualityComparer` (that checks the exact values of _numerator_, _denominator_ and _state_)
Are you sure this should be checking the `State` too?

>     * [ ]  Data type `Fraction` uses  `FractionComparer.ValueEquality` for `CompareTo(...)` method.
Unless you intend to add something other than the standard math comparisons - i don't see any reason to have it outside the class (there is no performance drawback when comparing normalized fractions here)

>     * [ ]  Add new method overload `Fraction.Equals(Fraction, FractionComparison)` (simular to [StringComparison](https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparison?view=net-8.0))
Can't people just use the `StrictComparer,Instance` instead? Are we going to add support for `tolerance`? `Comparison.Absolute`, `Comparison.Relative` ?

danm-de:
> > New class `FractionStrictEqualityComparer` (that checks the exact values of _numerator_, _denominator_ and _state_)
> 
> Are you sure this should be checking the `State` too?

Actually yes. I'm thinking of a third comparer - although I couldn't think of a name for the numerator and denominator test.

> Unless you intend to add something other than the standard math comparisons - i don't see any reason to have it outside the class (there is no performance drawback when comparing normalized fractions here)

This was an idea to share common code between Equals and CompareTo.

> > Add new method overload `Fraction.Equals(Fraction, FractionComparison)` (simular to [StringComparison](https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparison?view=net-8.0))
> 
> Can't people just use the `StrictComparer.Instance` instead? Are we going to add support for `tolerance`? `Comparison.Absolute`, `Comparison.Relative` ?

Another idea. If you think we don't need this, I'll drop it.

lipchev:
> > Unless you intend to add something other than the standard math comparisons - i don't see any reason to have it outside the class (there is no performance drawback when comparing normalized fractions here)
> 
> This was an idea to share common code between Equals and CompareTo.

That would hurt the performance of the Equals unnecessarily.

> > > Add new method overload `Fraction.Equals(Fraction, FractionComparison)` (simular to [StringComparison](https://learn.microsoft.com/en-us/dotnet/api/system.stringcomparison?view=net-8.0))
> > 
> > 
> > Can't people just use the `StrictComparer.Instance` instead? Are we going to add support for `tolerance`? `Comparison.Absolute`, `Comparison.Relative` ?
> 
> Another idea. If you think we don't need this, I'll drop it.

I don't mind it of course- just paranoid about having an extra enum to worry about (what states are you thinking about Default, Strict and _Stricter_ 😄 )?

danm-de:
> That would hurt the performance of the Equals unnecessarily.

Do you think so? I honestly don't know how the JIT behaves with `readonly struct`. Technically, only the numerator and denominator would have to be copied for the comparisons - all other functions are pure. Let's see if we can avoid calling the `Reduce()` method within `Equals`.

> I don't mind it of course- just paranoid about having an extra enum to worry about (what states are you thinking about Default, Strict and _Stricter_ 😄 )?

*ÜberStrict* of course. However, I spoke to a colleague - probably YAGNI. 

- ValueEquality -> Compare the actual value of $numerator / denominator$.
- Strict -> Compare the the exact values of `numerator` and `denominator` (same as `Fraction.Equals(..)` today). 

I discarded the idea of ​​implementing `IComparable` in `FractionComparer`. I wanted to use `StringComparer` as a template - it makes sense there as the sort order actually changes. But that doesn't apply to `Fraction`.

New default is ValueEquality.

lipchev:
> > That would hurt the performance of the Equals unnecessarily.
> 
> Do you think so? I honestly don't know how the JIT behaves with `readonly struct`. Technically, only the numerator and denominator would have to be copied for the comparisons - all other functions are pure. Let's see if we can avoid calling the `Reduce()` method within `Equals`.

I mean that the implementations are slightly different, where `CompareTo` ends up with `Sign.CompareTo(other.Sign)`, the other one does `Equals(..)` (which is faster) . There is also another shortcut that is also only applicable to the Equals method- but tell me when you're ready I'll have a stab at it.

I agree 100% with the rest.

danm-de:
> I mean that the implementations are slightly different, where `CompareTo` ends up with `Sign.CompareTo(other.Sign)`, the other one does `Equals(..)` (which is faster) . There is also another shortcut that is also only applicable to the Equals method- but tell me when you're ready I'll have a stab at it.

https://github.com/danm-de/Fractions/pull/66/commits/3e55140b2a0056cead5431dd72d50955d1cddbfb

danm-de:
> > I mean that the implementations are slightly different, where `CompareTo` ends up with `Sign.CompareTo(other.Sign)`, the other one does `Equals(..)` (which is faster) . There is also another shortcut that is also only applicable to the Equals method- but tell me when you're ready I'll have a stab at it.

I have no idea what I'm doing here... 🤡  https://github.com/danm-de/Fractions/commit/1d22ada078e5daf1f2ad37069ae6c3745226c839

danm-de:
I think that's it. Hopefully I haven't forgotten anything.

danm-de:
@lipchev 
Do you think we can dare to release version 8.0.0?

PR #68, opened 2024-05-26
fixing the comparisons with infinity 

lipchev:
- added extra tests for the `CompareTo` and the `Equals` methods
- updated the benchmark results (no difference w.r.t. `ref readonly` modifier)

### DIFF
--- a/src/Fractions/Fraction.CompareTo.cs
+++ b/src/Fractions/Fraction.CompareTo.cs
@@ -78,7 +78,7 @@ public readonly partial struct Fraction
         }
 
         if (denominator2.IsZero) {
-            return numerator2.Sign; // PositiveInfinity -> -1, NegativeInfinity -> 1
+            return -numerator2.Sign; // PositiveInfinity -> -1, NegativeInfinity -> 1
         }
 
         // both values are non-zero fractions with different denominators  
--- a/tests/Fractions.Tests/FractionSpecs/CompareTo/Method_CompareTo.cs
+++ b/tests/Fractions.Tests/FractionSpecs/CompareTo/Method_CompareTo.cs
@@ -1,5 +1,8 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Numerics;
+using FluentAssertions;
 using NUnit.Framework;
 using Tests.Fractions;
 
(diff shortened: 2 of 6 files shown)


lipchev:
@danm-de I cannot reopen the previous PR so here is another one. There are perhaps way too many tests but I've been using them to test every `return` statement on my optimized implementation. It's almost ready, and significantly faster- but I want to check a few more things before I publish it (probably tonight), so it would be easier if kept the extra tests for now..

tests/Fractions.Tests/FractionSpecs/CompareTo/Method_CompareTo.cs
```diff
@@ -103,6 +106,20 @@ public class When_comparing_two_fractions : Spec {
...
+            yield return new TestCaseData(new Fraction(-42, 66, false), Fraction.PositiveInfinity).Returns(-1);
+            yield return new TestCaseData(new Fraction(42, -66, false), Fraction.PositiveInfinity).Returns(-1);
+            yield return new TestCaseData(new Fraction(-42, -66, false), Fraction.PositiveInfinity).Returns(-1);
+            // Any number with NegativeInfinity
+            yield return new TestCaseData(Fraction.One, Fraction.NegativeInfinity).Returns(1);
+            yield return new TestCaseData(Fraction.MinusOne, Fraction.NegativeInfinity).Returns(1);
+            yield return new TestCaseData(new Fraction(42, 66, false), Fraction.NegativeInfinity).Returns(1);
+            yield return new TestCaseData(new Fraction(-42, 66, false), Fraction.NegativeInfinity).Returns(1);
+            yield return new TestCaseData(new Fraction(42, -66, false), Fraction.NegativeInfinity).Returns(1);
+            yield return new TestCaseData(new Fraction(-42, -66, false), Fraction.NegativeInfinity).Returns(1);
```
lipchev, line 122: These were all failing before..

src/Fractions/Fraction.CompareTo.cs
```diff
@@ -78,7 +78,7 @@ namespace Fractions;
         }
 
         if (denominator2.IsZero) {
-            return numerator2.Sign; // PositiveInfinity -> -1, NegativeInfinity -> 1
+            return -numerator2.Sign; // PositiveInfinity -> -1, NegativeInfinity -> 1
```
lipchev, line 81: I had to bring out `the big hammer` for this one.. 🔨

danm-de (approved):

PR #80, opened 2024-07-04
Optimizing the FromDoubleRounded (with significant digits)

lipchev:
- `ArgumentOutOfRangeException` is thrown when `significantDigits` is < 1 or > 17
- updating relevant tests
- updating the FromDoubleRoundedBenchmark (adding the `ReduceTerms` param)
- removed some obsolete Resources (`PositiveInfinityNotSupported` etc.)

This is actually a very important improvement (beyond the differences shown in the charts below) as in the previous implementation we get this:
<!DOCTYPE html>
<HTML>
<body>

  | Name | Value | Type
-- | -- | -- | --
◢ | Fraction.FromDoubleRounded(1.2, 15, false) | {120000000000000/100000000000000} | Fractions.Fraction

</body>
</HTML>

Which should now change to:
<!DOCTYPE html>
<HTML>
<body>

  | Name | Value | Type
-- | -- | -- | --
◢ | Fraction.FromDoubleRounded(1.2, 15, false) | {12/10} | Fractions.Fraction

</body>
</HTML>

### DIFF
--- a/benchmarks/Fractions.Benchmarks/FromDoubleBenchmarks.cs
+++ b/benchmarks/Fractions.Benchmarks/FromDoubleBenchmarks.cs
@@ -5,6 +5,7 @@ using BenchmarkDotNet.Attributes;
 namespace Fractions.Benchmarks;
 
 [MemoryDiagnoser]
+[ShortRunJob]
 public class FromDoubleBenchmarks {
     public FromDoubleBenchmarks() {
         // initialize static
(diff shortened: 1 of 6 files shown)


lipchev:
Here are the relevant benchmarks and values (with both `reduceTerms=True` and the `reduceTerms=False`):

image

And here are the allocation differences (_as % of base_):

image

And finally here are the absolute times for context (just the _interesting_ values):

image

danm-de (approved):

PR #81, opened 2024-07-04
Implemened the `INumber<Fraction>` interface (and more..)

### LINKED ISSUE #82: Fraction.ToDouble() returns NaN for {2 * MaxValue,  2 * MaxValue}
Let's illustrate this with an example:

`new Fraction((BigInteger)double.MaxValue * 2, (BigInteger)double.MaxValue * 2, false)`

<!DOCTYPE html>
<HTML>
<head>
<body>

  | Name | Value | Type
-- | -- | -- | --
▶ | A | 3.59538627e+308 | System.Numerics.BigInteger
▶ | B | 3.59538627e+308 | System.Numerics.BigInteger
  | State | Unknown | Fractions.FractionState
▶ | StringFormats | "1" | Fractions.Fraction.FractionDebugView.StringFormatsView
◢ | ValueFormats | NaN | Fractions.Fraction.FractionDebugView.NumericFormatsView
  | Decimal | 1 | decimal
  | Double | NaN | double
  | Integer | 1 | int
  | Long | 1 | long

</body>
</HTML>

I actually discovered the issue when one of my square roots turned up with a `0`:
```
        var largeNumber = BigInteger.Pow(10, 309);
        var fraction = new Fraction(4 * largeNumber, largeNumber, false);
        var actual = fraction.Sqrt();  // is Fraction.Zero..
```

..but I would have likely caught it when testing the `INumber` interface and the `TryConvertToChecked`, `TryConvertToSaturating` and `TryConvertToTruncating` (which are the only functions that require _special attention_).

I've already started implementing the `INumber<Fraction>` interface (two weeks ago) but wanted to finish up the tests for my (_still unpublished_) PR for UnitsNet - when one of the tests failed (_note this wasn't some static test concocted with an extreme value, it was an actual root-sum-of-squares that failed one of the tests in my own code-base_)..

lipchev:
@danm-de  _Sorry, I didn't realize what branch I was on, and pushed my commit to the branch of this PR.. I'll update the title and description, but if you prefer- I can split them later. Anyway, there aren't any stoppers for me here- the same fix can be applied on my side, until this is merged, so take your time._ 

## Implementation of the `INumber<Fraction>` interface

- introduced the `ToDecimalSaturating` method which saturates the output value to the decimal range [`MinValue`, `MaxValue`] 
- improved the accuracy (and performance) of the checked `ToDecimal()` when the terms are beyond the decimal range
- fixed the possible loss of precision of the `ToDouble()` method when one of the terms is exceeding the range of the double (fixes #82 )
- updated the tests and added some _extreme_ values the `ConvertToNumberBenchmark` (_no change among the previously tested values_)

### Fraction.Round: Adding a Fraction.Round overload with the "normalize" parameter:
- when set to true- it ensures that the result is normalized (which wasn't always the case before)
- when set to false- it maintains the decimal denominator (and skipping the GCD reduction)
- the DecimalNotationFormatter now uses the non-normalized version (which should in theory be more performant)
- added the required tests in Method_Round
- updated the DecimalNotationBenchmarkResults (_not much of a difference as shown by the benchmarks in the next comment_)


lipchev:
There wasn't actually much of a visible change on the DecimalNotationFormatterBenchmarks (to my surprise) - but I i expect the `Round` itself (which isn't benchmarked anywhere) would likely show the _normal_ difference..
Here are some of the cases for which I know (for sure) that we're hitting the _new_ round..

image

The rest of it looks like noise..

image

And the allocations have just one very small decrease...

image

tests/Fractions.Tests/FractionSpecs/ToDataType/Method_ToDataType.cs
```diff
@@ -143,6 +143,54 @@ public class When_converting_NaN : Spec {
...
+        ulong.CreateSaturating(Fraction.NaN).Should().Be(default);
+        int.CreateSaturating(Fraction.NaN).Should().Be(default);
+        uint.CreateSaturating(Fraction.NaN).Should().Be(default);
+        nint.CreateSaturating(Fraction.NaN).Should().Be(nint.Zero);
+        UIntPtr.CreateSaturating(Fraction.NaN).Should().Be(UIntPtr.Zero);
+        short.CreateSaturating(Fraction.NaN).Should().Be(default);
+        ushort.CreateSaturating(Fraction.NaN).Should().Be(default);
+        byte.CreateSaturating(Fraction.NaN).Should().Be(default);
+        sbyte.CreateSaturating(Fraction.NaN).Should().Be(default);
+    }
```
lipchev, line 192: Don't know what the reason is, but this is how this is implemented in BCL.. I first thought it was just in order to avoid throwing an exception, but later realized that `BigInteger.CreateSaturating(double.PositiveInfinity)` actually throws an `OverflowException`..

tests/Fractions.Tests/FractionSpecs/ToDataType/Method_ToDataType.cs
```diff
@@ -219,4 +351,442 @@ public class When_converting_NegativeInfinity : Spec {
...
+        ulong.CreateTruncating(fraction).Should().Be(18446744073709551615UL);
+        int.CreateTruncating(fraction).Should().Be(-1);
+        uint.CreateTruncating(fraction).Should().Be(4294967295u);
+        nint.CreateTruncating(fraction).Should().Be(-1);
+        UIntPtr.CreateTruncating(fraction).Should().Be(new UIntPtr(18446744073709551615));
+        short.CreateTruncating(fraction).Should().Be(-1);
+        ushort.CreateTruncating(fraction).Should().Be(65535);
+        byte.CreateTruncating(fraction).Should().Be(255);
+        sbyte.CreateTruncating(fraction).Should().Be(-1);
+    }
```
lipchev, line 790: This is another contentious point- currently all integer based types (including the unsigned versions) are created using an intermediate conversion to `BigInteger`:
- `TOther.CreateSaturating((BigInteger)fraction)  // where TOther might be the "int" type`
- `TOther.CreateTruncating((BigInteger)fraction) `

The `BigInteger` has separate implementation for each of the `TryConvertTo..` versions, with the truncating one, doing the overflowing as demonstrated by the second set of tests.
However, as I was looking at the sources for the `decimal`/`double` types, it seems they have just one implementation for the `TryConvertToSaturating` and `TryConvertToTruncating` (with the latter calling the former)..

I assume, whoever wants to create unsigned numbers from fractions knows what he's doing.. 🤷

src/Fractions/Fraction.ConvertTo.cs
```diff
@@ -81,33 +87,167 @@ namespace Fractions;
...
+    ///     Converts the fraction to a decimal number. If the fraction cannot be represented as a decimal due to its
+    ///     size, the method will return the closest possible decimal representation (either decimal.MaxValue or
+    ///     decimal.MinValue).
+    /// </summary>
+    /// <returns>
+    ///     The decimal representation of the fraction. If the fraction is too large to represent as a decimal, returns
+    ///     decimal.MaxValue. If the fraction is too small to represent as a decimal, returns decimal.MinValue.
+    /// </returns>
+    /// <remarks>If the fraction is NaN (the numerator and the denominator are both zero), the method will return decimal.Zero.</remarks>
+    public decimal ToDecimalSaturating() {
```
lipchev, line 164: Of course this one doesn't have to public, but I guess somebody might find some use for it (also, it was easier to test, rather than having to go through the INumber interface... 😄  ).

As a side note- a lot of the methods from `INumber` that are redundant with our existing properties (or are generally not interesting to the user) were implemented as explicit. Feel free to add or remove the methods that you feel would/ wouldn't be useful to the user (I'm talking about things like `IsFinite(Fraction)`, `IsInteger(Fraction)` etc..).

lipchev:
I've created a new pull request for the first part (the original pull request) for the rounding overload : #95
I'll close this and create one (or possibly two) other pull requests regarding the `INumber` interface shortly..

PR #89, opened 2024-11-19
DecimalNotationFormatter: extracted a non-boxing Format overload

lipchev:
I would normally put the Fraction as the first argument, but the current overload would be picked up automatically for anyone that's already using something of the form:

`DecimalNotationFormatter.Instance.Format(format, _fraction, formatProvider);`

..provided that the `_fraction` is of type `Fraction`.

I could of course swap them if you prefer..

### DIFF
--- a/src/Fractions/Formatter/DecimalNotationFormatter.cs
+++ b/src/Fractions/Formatter/DecimalNotationFormatter.cs
@@ -76,11 +76,27 @@ public class DecimalNotationFormatter : ICustomFormatter {
 
     private static readonly BigInteger Ten = new(10);
 
-    /// <summary>
-    ///     Formats the value of the specified Fraction object as a string using the specified format.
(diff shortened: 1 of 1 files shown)


danm-de:
Thanks!

I think using the same signature as defined in `ICustomFormatter` is okay.

lipchev:
> I would normally put the Fraction as the first argument, but the current overload would be picked up automatically for anyone that's already using something of the form:
> 
> `DecimalNotationFormatter.Instance.Format(format, _fraction, formatProvider);`
> 
> ..provided that the `_fraction` is of type `Fraction`.

..this would have been true if I hadn't made the overload static 😆 .. Well, personally I'm fine with using `DecimalNotationFormatter.Format(..)`

Alternatively we could have an [Obsolete] instance method forwarding to the static one..

PR #90, opened 2024-11-21
Add net9.0 to Fractions.csproj

lipchev:
- [x] Updated the net8.0 target to net9.0 for the Fractions.Benchmarks.csproj and update all existing jobs
- [x] Posting all benchmark results (see the graphs in comment below)

There appear to be quite a few interesting performance improvements coming with the net9.0 release, here are a few that caught my eye:

- https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-9/#biginteger
- https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-9/#primitive-types
- https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-9/#object-stack-allocation

Great stuff 👏


benchmarks/Fractions.Benchmarks/Fractions.Benchmarks.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFrameworks>net481;net8.0</TargetFrameworks>
+    <TargetFrameworks>net9.0;net48</TargetFrameworks>
```
lipchev, line 5: I think there was thing in past about having the framework project first, but it seems to be fine now...

benchmarks/Fractions.Benchmarks/Fractions.Benchmarks.csproj
```diff
@@ -12,14 +12,6 @@
     <PackageReference Include="BenchmarkDotNet" Version="0.14.0" />
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3">
-      <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-    </PackageReference>
-    <PackageReference Include="System.Reflection.Metadata" Version="9.0.0" />
-  </ItemGroup>
```
lipchev, line 21: I don't think this is currently necessary: I'm getting some binding redirect warnings for `System.Collections.Immutable` and the like, but nothing too concerning (hopefully we get an update from BDN with the new packages that came out last week).

lipchev:
And here are the goods (just the highlights):

image

image

image

image

image

image

image

image

image

image

Note that the previous results correspond to the difference from the benchmark results that I updated yesterday (incl. the optimization)... however if we bring things into their absolute scale, the difference well.. I wouldn't call it a "game changer" 😄 

image

image

image

image

For the last two, the comparison is against the "boxing" version of the `DecimalNotationFormatter` (I didn't bother to update it yesterday):

image

image

benchmarks/Fractions.Benchmarks/Fractions.Benchmarks.csproj
danm-de: Unfortunately, this change doesn't make sense to me - nothing changes in the compiler result.

benchmarks/Fractions.Benchmarks/Fractions.Benchmarks.csproj
danm-de: I fixed exactly these build warnings with it. See https://learn.microsoft.com/en-us/dotnet/framework/migration-guide/reference-assemblies

benchmarks/Fractions.Benchmarks/Fractions.Benchmarks.csproj
lipchev: I'm not sure of all the things that change in regards to the whole process, but if we put the framework first it uses it for the "host" and outputs the benchmark results in a `net481` folder instead of the `net9.0` folder where I would expect them to be. I think this also affects the nuget packages it's using which cause the warning (the auto-generated projects seem to be fine as far as i can tell).

benchmarks/Fractions.Benchmarks/Fractions.Benchmarks.csproj
lipchev: I don't know, maybe it's because I already have the SDK installed and it's not giving me trouble- are you still seeing the warnings after swapping the targets?

danm-de: This is indeed unexpected behavior. But it's not really surprising when you consider the possibilities of Roslyn compiler extensions.

lipchev:
@danm-de Any chance we could revive this?

PR #94, opened 2024-12-10
DecimalNotationFormatter no longer throws when failing to parse the format specifier

### LINKED ISSUE #93: DecimalNotationFormatter throws FormatException for some custom formats
Not sure why I added the [this validation](https://github.com/danm-de/Fractions/blob/e04ce0fc4101c15704b4e5ef575a58cc2a4099f1/src/Fractions/Formatter/DecimalNotationFormatter.cs#L243), but I think it was a mistake: 

<!DOCTYPE html>
<HTML>
<head>
<body>

  | Name | Value | Type
-- | -- | -- | --
  | DecimalNotationFormatter.Format("Test: 0.00", 123.45m, null) | "Test: 123.45" | string
  | 123.45m.ToString("Pest: 0.00") | "Pest: 123.45" | string 
◢ | DecimalNotationFormatter.Format("Pest: 0.00", 123.45m, null) | 'DecimalNotationFormatter.Format("Pest: 0.00", 123.45m, null)' threw an exception of type 'System.FormatException' | string {System.FormatException}
  | 123.45.ToString("P-1") | "P-1" | string

</body>
</HTML>

Neither the double or the decimal equivalents throw an exception..

lipchev:
… (fixes #93)

### DIFF
--- a/src/Fractions/Formatter/DecimalNotationFormatter.cs
+++ b/src/Fractions/Formatter/DecimalNotationFormatter.cs
@@ -235,25 +235,23 @@ public class DecimalNotationFormatter : ICustomFormatter {
         };
     }
 
-
-    private static int GetPrecisionDigitsOrDefault(string format, int defaultValue) {
-        return format.Length > 1 ? GetPrecisionDigits(format) : defaultValue;
-    }
-
-    private static int GetPrecisionDigits(string format) {
- #if NET
-        if (!int.TryParse(format.AsSpan(1), out var nbDecimals) || nbDecimals < 0) {
-            throw new FormatException($"The {format} format string is not supported.");
-        }
+    private static bool TryGetPrecisionDigits(string format, int defaultPrecision, out int maxNbDecimals) {
+        if (format.Length != 1) {
+#if NET
+            if (!int.TryParse(format.AsSpan(1), out maxNbDecimals)) {
+                return false;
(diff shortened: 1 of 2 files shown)


src/Fractions/Formatter/DecimalNotationFormatter.cs
```diff
@@ -235,25 +235,23 @@ public static string Format(string? format, Fraction fraction, IFormatProvider?
...
+            }
 #endif
+        }
 
-        return nbDecimals;
+        maxNbDecimals = defaultPrecision;
+        return true;
     }
-
+
```
danm-de, line 254: I've added some changes.
image


```suggestion
        private static bool TryGetPrecisionDigits(string format, int defaultPrecision, out int maxNbDecimals) {
        if (format.Length == 1) {
            // The number of digits is not specified, use default precision.
            maxNbDecimals = defaultPrecision;
            return true;
        }

#if NET
        if (int.TryParse(format.AsSpan(1), out maxNbDecimals)) {
            return true;
        }
#else
        if (int.TryParse(format.Substring(1), out maxNbDecimals)) {
            return true;
        }
#endif

        // Seems to be some kind of custom format we do not understand, fallback to default precision.
        maxNbDecimals = defaultPrecision;
        return false;
    }
```

PR #95, opened 2024-12-21
Adding a Fraction.Round overload with the "normalize" parameter

lipchev:
- when set to true- it ensures that the result is normalized (which wasn't always the case before)
- when set to false- it maintains the decimal denominator (and skipping the GCD reduction)
- the `DecimalNotationFormatter` now uses the non-normalized version (which is more performant)
- added the required tests in Method_Round


danm-de:
Merged, thank you!

PR #96, opened 2024-12-21
Implemented the INumber<Fraction> interface:

lipchev:
- introduced the `ToDecimalSaturating` method which saturates the output value to the decimal range `[MinValue, MaxValue]`
- `ToDecimal` now also throws an `OverflowException` for values that are very close to `0` (**_breaking change_**)
- improved the accuracy (and performance) of the checked `ToDecimal` when the terms are beyond the decimal range
- fixed the possible loss of precision of the existing `ToDouble` when one of the terms is exceeding the range of the double (#82)
- completed the coverage of the `CompareTo` methods (100%)
- improved the coverage of the `Add` method (94%)

There is still quite a few tests that are missing, but for the most part- I think it should be bug free 😛 

I was thinking of splitting this into 2-3 PRs: one introducing the fixes for the `ToDouble` / `ToDecimal` - which now have 100% coverage) and another one for the introduction of the "++" / "--" operators, however I'm not going to make it tonight...If time permits, and if you haven't got any remarks I'll try to do it later..


danm-de:
Thanks for your work!

I'll take a closer look at your pull request once my new PC arrives. I made a big mistake with my personal Surface Book 2 and upgraded to Windows 11. Since then, it has hardly been possible to work properly in Visual Studio.

I'm not entirely sure how we do this with the license from the Fraction project. I would be happy if you create a contributor file and document yourself there (listing your changes). I changed the project's license from BSD-2-clause to MIT to make it easier to fork on GitHub etc. The FluentAssertion debacle has caused me great concern and I want to make it clear that this project will never be commercialized in this way. It will always be freely (free) available to everyone – whether private or business.

lipchev:
> I'll take a closer look at your pull request once my new PC arrives. I made a big mistake with my personal Surface Book 2 and upgraded to Windows 11. Since then, it has hardly been possible to work properly in Visual Studio.

That's a relief- frankly I was worried that you might have gotten yourself into a skiing accident or something.. 😌 
 
> I changed the project's license from BSD-2-clause to MIT to make it easier to fork on GitHub etc. The FluentAssertion debacle has caused me great concern and I want to make it clear that this project will never be commercialized in this way. It will always be freely (free of costs) available to everyone – whether private or business.

Glad to hear it, I feel the same way..

> I'm not entirely sure how we do this with the license from the Fraction project. I would be happy if you create a contributor file and document yourself there (listing your changes). 

Frankly, if that's not part of the licensing requirements, I really don't care- and if it requires me to document more stuff, I really rather go without the attributions.. Besides, I'm pretty sure the future AIs would be able to figure these things out on their own.. 😚

tests/Fractions.Tests/FractionSpecs/ToDataType/Method_ToDataType.cs
```diff
@@ -220,3 +235,207 @@ public void ToDouble_should_return_NegativeInfinity() {
...
+            .Should()
+            .Throw<OverflowException>();
+    }
+    
+    [Test]
+    public void An_OverflowException_is_thrown_when_result_is_smaller_than_MinValue() {
+        Invoking(() => (2 * new Fraction(decimal.MinValue)).ToDecimal())
+            .Should()
+            .Throw<OverflowException>();
+    }
```
lipchev, line 366: I just realized that I had missed two tests here:

``` csharp
    [Test]
    public void An_OverflowException_is_thrown_when_result_is_positive_and_smaller_than_One_over_MaxValue() {
        Invoking(() => new Fraction(1, 2 * (BigInteger)decimal.MaxValue).ToDecimal())
            .Should()
            .Throw<OverflowException>();
    }

    [Test]
    public void An_OverflowException_is_thrown_when_result_is_negative_and_larger_than_MinusOne_over_MaxValue() {
        Invoking(() => new Fraction(-1, 2 * (BigInteger)decimal.MaxValue).ToDecimal())
            .Should()
            .Throw<OverflowException>();
    }
```

and surprise, surprise - they ended up failing... The fix is easy- I just remove two lines of code and everything is green again, however this would be a breaking change.

Are you planning a version bump for this PR? And of course, are you Ok with the change (note that this would affect the behavior of the explicit cast).

tests/Fractions.Tests/FractionSpecs/NumberInterface/Method_IsRealNumber.cs
lipchev, line 1: I've started a new folder with specs for the INumber interfaces- with `#if NET` in the file- we could also exclude this from csproj, but when I last tried this (an year or so ago) it wasn't behaving nicely (the file didn't immediately appear when switching between the frameworks).

tests/Fractions.Tests/FractionSpecs/ToDataType/Method_ToDataType.cs
danm-de: I think that the introduction of the INumber interface is a major change. In fact, the effects (apart from those described in the documentation) are not quite clear to me yet - maybe some framework functions will behave differently afterwards (Convert, LINQ, ToStringFormatter, ???).

Of course, the major version bump also offers room for further breaking changes.

lipchev: I'm pretty sure that the string interpolation makes use of the `ISpanFormattable`, I don't know of anything else that works out of the box, however the interface provides _the opportunity_ for many extension methods to be implemented.. One notable example are the extensions for `Sum`, `Average` etc. You'd think that they'd be the first ones to get added- but no.. (see https://github.com/dotnet/runtime/issues/64031).

lipchev: See what we have in UnitsNet:  https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/GenericMath/GenericMathExtensions.cs or the more constrained version https://github.com/angularsen/UnitsNet/blob/release/v6/UnitsNet/GenericMath/GenericMathExtensions.cs (there is an even funkier version that's still unpublished, where we implement `IAdditionOperators<Temperature, TemperatureDelta, Temperature>` but I don't have a for it just yet..)

lipchev: fixed in https://github.com/danm-de/Fractions/pull/96/commits/034b9f376b8739a379cbf5090c99f220b04c8634

lipchev:
@danm-de Please consider whether we want to have the `CreateChecked`, `CreateSaturating` and `CreateTruncating` functions explicit or not. The standard numeric types have each of them as a regular function (overriding the default interface implementations). Here's what this could look like:

``` csharp
    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    public static Fraction CreateChecked<TOther>(TOther value) where TOther : INumberBase<TOther>
    {
        Fraction result;
        if (typeof(TOther) == typeof(Fraction))
            result = (Fraction )(object)value;
        else if (!TryConvertChecked(value, out result) && !TOther.TryConvertToChecked(value, out result))
            throw new NotSupportedException($"Cannot create a {nameof(Fraction)} from the given value of type {typeof(TOther)}.");
        return result;

        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        static bool TryConvertChecked<TSource, TTarget>(TSource source, out TTarget? result)
            where TSource : INumberBase<TSource>
            where TTarget : INumberBase<TTarget>
        {
            return TTarget.TryConvertFromChecked(source, out result);
        }
    }
```

It's kinda silly looking (x3) but there's no other way of calling the explicit implementation of the `TryConvertFrom` method. On the other hand, extracting it's implementation in a non-explicit method, makes the explicit version _untestable_.

src/Fractions/Fraction.cs
```diff
@@ -146,4 +153,180 @@ namespace Fractions;
...
+
+        if (numerator.IsZero) {
+            return denominator.IsZero; // if we want to consider NaN as "canonical"
+        }
+
+        if (denominator.IsZero) {
+            return numerator.IsOne || numerator == BigInteger.MinusOne;
+        }
+
+        return BigInteger.GreatestCommonDivisor(numerator, denominator).IsOne;
```
danm-de, line 276: Can we optimize this here, or am I missing something?

```suggestion
        return !value._normalizationNotApplied ||
               BigInteger.GreatestCommonDivisor(numerator, denominator).IsOne;
```

src/Fractions/Fraction.cs
```diff
@@ -146,4 +153,180 @@ namespace Fractions;
...
+    }
+
+    /// <summary>Determines if a value represents an even integral number.</summary>
+    /// <param name="value">The value to be checked.</param>
+    /// <returns>
+    ///     <see langword="true" /> if <paramref name="value" /> is an even integer; otherwise, <see langword="false" />.
+    /// </returns>
+    public static bool IsEvenInteger(Fraction value) {
+        var fraction = value.Reduce();
+        return fraction.Denominator.IsOne && fraction.Numerator.IsEven;
```
danm-de, line 286: I couldn't find any unit tests for this. What do you think about the following optimization?

```suggestion
        if (value.Denominator.IsZero) {
            return false;
        }

        if (!value._normalizationNotApplied) {
            return BigInteger.Abs(value.Denominator).IsOne && value.Numerator.IsEven;
        }

        var intValue = BigInteger.DivRem(value.Numerator, value.Denominator, out var remainder);
        return remainder.IsZero && intValue.IsEven;
```

src/Fractions/Fraction.cs
```diff
@@ -146,4 +153,180 @@ namespace Fractions;
...
+    }
+
+    /// <summary>Determines if a value represents an odd integral number.</summary>
+    /// <param name="value">The value to be checked.</param>
+    /// <returns>
+    ///     <see langword="true" /> if <paramref name="value" /> is an odd integer; otherwise, <see langword="false" />.
+    /// </returns>
+    public static bool IsOddInteger(Fraction value) {
+        var fraction = value.Reduce();
+        return fraction.Denominator.IsOne && !fraction.Numerator.IsEven;
```
danm-de, line 328: Can we optimize the method a little bit with this?

```suggestion
        if (value.Denominator.IsZero) {
            return false;
        }
        if (!value._normalizationNotApplied) {
            return BigInteger.Abs(value.Denominator).IsOne && !value.Numerator.IsEven;
        }
        var intValue = BigInteger.DivRem(value.Numerator, value.Denominator, out var remainder);
        return remainder.IsZero && !intValue.IsEven;
```

danm-de (changes requested):
Wow, you had a lot of work implementing the INumber interface.

src/Fractions/Fraction.ConvertTo-netcore.cs
```diff
@@ -0,0 +1,395 @@
...
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    static bool INumberBase<Fraction>.TryConvertToSaturating<TOther>(Fraction value, [MaybeNullWhen(false)] out TOther result) {
+        if (typeof(TOther) == typeof(decimal)) {
+            var convertedValue = value.ToDecimalSaturating();
+            result = (TOther)(object)convertedValue;
+            return true;
+        }
+
+        if (typeof(TOther) == typeof(double)) {
+            var convertedValue = value.ToDouble();
```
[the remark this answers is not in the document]
lipchev: Yes, I was also surprised by this one, but this is how the `TryConvertToSaturating` is implemented for the `BigInteger`:
```  csharp
            else if (typeof(TOther) == typeof(double))
            {
                double actualResult = (double)value;
                result = (TOther)(object)actualResult;
                return true;
            }
```

I'm not 100% sure of the reasoning, but my guess is that they consider `PositiveInfinity` (being _larger_ than `double.MaxValue`) to be better representation for an overflowing `BigInteger`.

src/Fractions/Fraction.ConvertTo-netcore.cs
```diff
@@ -0,0 +1,395 @@
...
+                    -1 => (TOther)(object)TNumber.MinValue,
+                    _ => default
+                };
+
+            return true;
+        }
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    static bool INumberBase<Fraction>.TryConvertToTruncating<TOther>(Fraction value, [MaybeNullWhen(false)] out TOther result) {
```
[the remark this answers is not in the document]
lipchev: Yeah, my guess is that these are only useful to a very small number of people (dealing with hashing or something).

My implementation is based on that of the `BigInteger`- if anyone knows better, let them come forward with a PR 😄

src/Fractions/Fraction.cs
lipchev: I'm looking at my other project and that's exactly how I've implemented it there. I've even left myself a little comment:

``` csharp
        // TODO see about first checking the magnitudes
        var quotient = BigInteger.DivRem(numerator, denominator, out var remainder);
        return remainder.IsZero && quotient.IsEven;
```

src/Fractions/Fraction.ConvertTo.cs
```diff
@@ -81,33 +83,163 @@ public BigInteger ToBigInteger() {
     public decimal ToDecimal() {
         var numerator = Numerator;
         var denominator = Denominator;
-        if (denominator.IsZero) {
-            throw new DivideByZeroException();
+        switch (denominator.Sign) {
+            case 0: {
+                throw new DivideByZeroException();
```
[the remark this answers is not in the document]
lipchev: I wasn't sure whether it's worth the breaking change- my guess is that for the `ToDecimal` people are likely using something like

```  csharp
try
{
  var decimalValue = fraction.ToDecimal();
  // ..
}
catch (DivisionByZero)
{
 //..
}
catch (OverflowException)
{
// ..
}
```

which will still catch the `NaN` or `Infinity` but if we made the same modification on the `ToBigInteger` then the _likely_ `try-catch` scenario will no longer catch the exception:
```  csharp
try
{
  var bigInteger = fraction.ToBigInteger();
  // ..
}
catch (DivisionByZero)
{
 //..
}
```

Of course this is probably not a very likely scenario, but still.. Should i change it (everywhere)?

src/Fractions/Fraction.ConvertTo.cs
```diff
@@ -81,33 +83,163 @@ public BigInteger ToBigInteger() {
...
+            var withoutDecimalPlaces = (decimal)BigInteger.DivRem(numerator, denominator, out var remainder);
+            return remainder.IsZero ? withoutDecimalPlaces : withoutDecimalPlaces + (decimal)remainder / (decimal)denominator;
+        }
+
+        if (denominator > MAX_DECIMAL) {
+            // since both terms are non-zero and the numerator is smaller (in magnitude) to the denominator
+            // the resulting number would be in the range [-1,1] (exclusive)
+            // we want to flip the operation: x = a/b -> 1/x = b/a
+            var decimalPart = BigInteger.DivRem(denominator, numerator, out var remainder);
+            return remainder.IsZero ? 1m / (decimal)decimalPart : 1m / ((decimal)decimalPart + (decimal)remainder / (decimal)numerator);
```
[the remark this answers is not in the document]
lipchev: Fixed in https://github.com/Fractions-DotNet/Fractions/pull/96/commits/9d8553118c56fbbcd419ac945abffcac7a10f47b

src/Fractions/Fraction.ConvertTo.cs
```diff
@@ -81,33 +83,163 @@ public BigInteger ToBigInteger() {
...
-        var lowPart = remainder * BigInteger.Pow(TEN, 28) / denominator;
-        var decimalPlaces = (decimal)lowPart / (decimal)Math.Pow(10, 28);
+            var withoutDecimalPlaces = BigInteger.DivRem(numerator, denominator, out var remainder);
+            if (withoutDecimalPlaces > MAX_DECIMAL) {
+                return decimal.MaxValue;
+            }
 
-        return withoutDecimalPlaces + decimalPlaces;
-    }
+            return remainder.IsZero ? (decimal)withoutDecimalPlaces : (decimal)withoutDecimalPlaces + (decimal)remainder / (decimal)denominator;
```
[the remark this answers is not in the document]
lipchev: Fixed in https://github.com/Fractions-DotNet/Fractions/pull/96/commits/9d8553118c56fbbcd419ac945abffcac7a10f47b

src/Fractions/Fraction.ConvertTo.cs
```diff
@@ -81,33 +83,163 @@ public BigInteger ToBigInteger() {
...
+                    return (decimal)numerator / (decimal)denominator;
+                }
+            }
+
+            var withoutDecimalPlaces = BigInteger.DivRem(numerator, denominator, out var remainder);
+            if (withoutDecimalPlaces < MIN_DECIMAL) {
+                return decimal.MinValue;
+            }
+
+            return remainder.IsZero ? (decimal)withoutDecimalPlaces : (decimal)withoutDecimalPlaces + (decimal)remainder / (decimal)denominator;
```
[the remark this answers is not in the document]
lipchev: Fixed in https://github.com/Fractions-DotNet/Fractions/pull/96/commits/9d8553118c56fbbcd419ac945abffcac7a10f47b

src/Fractions/Fraction.ConvertTo.cs
```diff
@@ -81,33 +83,163 @@ public BigInteger ToBigInteger() {
...
+        if (denominator > MAX_DECIMAL) {
+            // since both terms are non-zero and the numerator is smaller (in magnitude) to the denominator
+            // the resulting number would be in the range [-1,1] (exclusive)
+            // we want to flip the operation: x = a/b -> 1/x = b/a
+            var decimalPart = BigInteger.DivRem(denominator, numerator, out var remainder);
+            if (decimalPart < MIN_DECIMAL || decimalPart > MAX_DECIMAL) {
+                return decimal.Zero;
+            }
+
+            return remainder.IsZero ? 1m / (decimal)decimalPart : 1m / ((decimal)decimalPart + (decimal)remainder / (decimal)numerator);
```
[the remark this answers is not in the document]
lipchev: Fixed in https://github.com/Fractions-DotNet/Fractions/pull/96/commits/9d8553118c56fbbcd419ac945abffcac7a10f47b

tests/Fractions.Tests/FractionSpecs/Magnitude/Method_MaxMagnitude.cs
```diff
@@ -0,0 +1,54 @@
...
+        new object[] { Fraction.NegativeInfinity, Fraction.PositiveInfinity, Fraction.PositiveInfinity }
+    ];
+
+    [TestCaseSource(nameof(MaxMagnitudeTestCases))]
+    public void MaxMagnitude_ShouldReturnCorrectResult(Fraction x, Fraction y, Fraction expected) {
+#if NET
+        double.MaxMagnitude(x.ToDouble(), y.ToDouble()).Should().Be(expected.ToDouble());
+#endif
+        Fraction.MaxMagnitude(x, y).Should().Be(expected);
+    }
```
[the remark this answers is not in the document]
lipchev: I agree that as long as everything works, this is cleaner. However, in my experience, as soon as one of the two tests pops up a red label- it becomes much harder to correlate the test values.

src/Fractions/Fraction.cs
```diff
@@ -146,4 +153,180 @@ namespace Fractions;
     /// The fraction's state.
     /// </summary>
     public FractionState State => _normalizationNotApplied ? FractionState.Unknown : FractionState.IsNormalized;
+
+
+    #region INumber definitions
+
+#if NET7_0_OR_GREATER
```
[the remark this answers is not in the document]
lipchev: fixed in https://github.com/Fractions-DotNet/Fractions/pull/96/commits/a48dcb56b107395595fb83475a954edf09f43ad9

src/Fractions/Fraction.cs
lipchev: fixed in https://github.com/Fractions-DotNet/Fractions/pull/96/commits/1643913092e03ca145a91e4d5de95948f16949da

src/Fractions/Fraction.cs
lipchev: fixed in https://github.com/Fractions-DotNet/Fractions/pull/96/commits/1643913092e03ca145a91e4d5de95948f16949da

src/Fractions/Fraction.cs
lipchev: fixed in https://github.com/Fractions-DotNet/Fractions/pull/96/commits/1643913092e03ca145a91e4d5de95948f16949da

src/Fractions/Fraction.ConvertTo.cs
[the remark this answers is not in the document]
danm-de: Personally, I wouldn't expect a `DivisionByZeroException` to occur during a type conversion. At least not anymore since we support _NaN_ and _+/-Infinity_.

image

I certainly didn't take into account in the first version that this type of calculation (`x / 0`) takes place within the Convert-methods, and therefore this type of exception is thrown. The usual exceptions for type conversions are `OverflowException`, `NotSupportedException`, etc.

For me, this breaking change would therefore only be logical.

image

danm-de (approved):

PR #99, opened 2025-01-29
Re-implement the Sqrt function (fixes #97):

### LINKED ISSUE #97: Square root accuracy is ignored

lipchev:
Re-implement the `Sqrt` function (fixes #97):
- implementation based on https://github.com/SunsetQuest/NewtonPlus-Fast-BigInteger-and-BigFloat-Square-Root
- the accuracy parameter switched from "number of decimals" to "number of digits"
- completed the test coverage and added additional benchmarks

duplicate of #98


lipchev:
I ran the benchmarks against the previously suggested fix (from #97) and this is ~5-10x faster, but more importantly here the resulting terms are much closer to the desired accuracy than before (no more than 20% over the target), which means that we don't need to bother with rounding the output. 

Here's a comparison chart:

image

danm-de:
That's pretty cool! Do we need to include a copy of the original license? Or has the algorithm been modified in significant parts so that it is considered an independent work?

lipchev:
> That's pretty cool! Do we need to include a copy of the original license? Or has the algorithm been modified in significant parts so that it is considered an independent work?

No, the relevant parts are mostly copy-pasted- I've only cleaned up the names and removed a few bits where the resulting `BigInteger` was being rounded down (instead we're simply returning extra digits).

I wasn't sure whether to include a copy of the license on the file-level (in my other project there is already an auto-generated header) - the AI said that using method-level-attributions is fine, but who knows...

danm-de:
> I wasn't sure whether to include a copy of the license on the file-level (in my other project there is already an auto-generated header) - the AI said that using method-level-attributions is fine, but who knows...

Okay, let's ask to be sure (https://github.com/SunsetQuest/NewtonPlus-Fast-BigInteger-and-BigFloat-Square-Root/issues/2).

PR #108, opened 2025-03-16
FromDoubleRounded: replaced the usages of BigInteger.Pow(TEN, exponent)

lipchev:
- `FromDoubleRounded`: replaced the usages of BigInteger.Pow(TEN, exponent) with the new `PowerOfTen(exponent)`
- completed the test coverage of the `FromDoubleRounded` method
- updated the benchmark results

### DIFF
--- a/benchmarks/results/Fractions.Benchmarks.FromDoubleBenchmarks-report-github.md
+++ b/benchmarks/results/Fractions.Benchmarks.FromDoubleBenchmarks-report-github.md
@@ -1,10 +1,10 @@
 ```
 
-BenchmarkDotNet v0.14.0, Windows 10 (10.0.19045.5131/22H2/2022Update)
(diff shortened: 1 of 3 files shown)


lipchev:
This one is fairly straightforward improvement.. Here is the relative difference:

image

And the allocations are down by up to 50%:

image

For reference, here are the new absolute values (I'm not sure there is anything to be done about the case of `double.MaxValue`..):

image

And here are the other `FromDouble` methods (reduced, excluding the `double.MaxValue`):

image

Note that we're currently using the `FromDouble` (reducing) as the _default_ conversion function (e.g. explicit casting, checked conversions etc..):

Things kinda even out when we skip the term-reduction:

image

However I personally don't see any benefit in using that version (or the approximating alternative) - in fact I think they go completely against the idea of representing _exact decimal values_ (and would *strongly* advice against using either of them: using `FromDoubleRounded(value, 15)` is the way to go (_or it's non-reducing alternative_).

PR #109, opened 2025-03-16
FromDecimal: replaced the usage of BigInteger.Pow(TEN, exponent)

lipchev:
- `FromDecimal`: replaced the usage of `BigInteger.Pow(TEN, exponent)`
- updated the benchmark results


lipchev:
As expected, this reduces the time by up to 40%:

image

And the allocations by up to 50%:

image

Here are the absolute times for constructors with `double` and `decimal` (both reducing):

image

PR #110, opened 2025-03-16
Optimizing the PowerOfTen function for exponents in the range [18, 54]

lipchev:
- Optimizing the `PowerOfTen(exponent)` function for exponents in the range [18, 54]
- increasing the number of powers tested in the the Rounding benchmarks
- updating the benchmark results (including the `[MethodImpl(MethodImplOptions.AggressiveInlining)]` which was previously missing)


src/Fractions/Fraction.Math.cs
lipchev, line 1: I was initially going to suggest that we make this function public. Even had a unit test for it:
``` csharp
    [Test]
    public void PowerOfTen_WithLargeExponent_Equals_BigIntegerPow([Values(18, 19, 36, 37, 54, 55)] int power) {
        var powerOfTen = Fraction.PowerOfTen(power);
        var bigIntegerPower = BigInteger.Pow(10, power);
        Assert.That(powerOfTen, Is.EqualTo(bigIntegerPower));
    }
```
However, after thinking about the use cases for it, decided to suggest something different. Here's what I ended up doing in my other project:

```  csharp
    /// <summary>
    ///     Creates a new instance of the <see cref="QuantityValue" /> struct with the specified numerator, denominator,
    ///     and a power of ten exponent.
    /// </summary>
    /// <param name="numerator">The numerator of the fraction representing the value.</param>
    /// <param name="denominator">The denominator of the fraction representing the value.</param>
    /// <param name="powerOfTen">
    ///     The power of ten to adjust the value. A positive exponent multiplies the numerator by 10 raised to the power,
    ///     while a negative exponent multiplies the denominator by 10 raised to the absolute value of the power.
    /// </param>
    /// <returns>
    ///     A new <see cref="QuantityValue" /> instance representing the specified value, corresponding to the scientific
    ///     notation: (numerator/denominator) * 10 ^ powerOfTen.
    /// </returns>
    /// <remarks>
    ///     This constructor enables precise representation of values using a fraction and a power of ten, 
    ///     corresponding to the scientific notation: (numerator/denominator) * 10 ^ powerOfTen.
    /// </remarks>
    public static QuantityValue FromPowerOfTen(BigInteger numerator, BigInteger denominator, int powerOfTen);
```

There is no `reduceTerms` parameter, as the `QuantityValue` doesn't bother with such things, but if you want to I can add it with a corresponding parameter.

This could as well be a constructor if you prefer.

src/Fractions/Fraction.Math.cs
```diff
@@ -555,10 +555,30 @@ public static Fraction Pow(Fraction @base, int exponent) {
...
+            result *= maxPowerOfTen;
+            exponent -= maxExponent; // the exponent is no more than maxExponent (18)
+        }
+
+        if (exponent == maxExponent) {
+            return result * maxPowerOfTen;
+        }
+
+        return result * PowersOfTen[exponent];
+    }
```
lipchev, line 581: I did some tests, and it appears that we start loosing after the 4th iteration (technically we're still faster, but at the cost of more allocations):

<html><body>


Method | Job | Runtime | Power | Mean | Error | StdDev | Gen0 | Allocated
-- | -- | -- | -- | -- | -- | -- | -- | --
PowerOfTen | .NET 9.0 | .NET 9.0 | 36 | 24.66 ns | 0.077 ns | 0.072 ns | 0.0024 | 40 B
BigIntegerPow | .NET 9.0 | .NET 9.0 | 36 | 83.95 ns | 0.391 ns | 0.366 ns | 0.0024 | 40 B
PowerOfTen | .NET Framework 4.8 | .NET Framework 4.8 | 36 | 50.15 ns | 0.243 ns | 0.227 ns | 0.0063 | 40 B
BigIntegerPow | .NET Framework 4.8 | .NET Framework 4.8 | 36 | 102.70 ns | 0.118 ns | 0.092 ns | 0.0191 | 120 B
PowerOfTen | .NET 9.0 | .NET 9.0 | 37 | 38.84 ns | 0.197 ns | 0.185 ns | 0.0048 | 80 B
BigIntegerPow | .NET 9.0 | .NET 9.0 | 37 | 89.08 ns | 0.131 ns | 0.122 ns | 0.0024 | 40 B
PowerOfTen | .NET Framework 4.8 | .NET Framework 4.8 | 37 | 79.30 ns | 0.077 ns | 0.064 ns | 0.0139 | 88 B
BigIntegerPow | .NET Framework 4.8 | .NET Framework 4.8 | 37 | 107.46 ns | 0.083 ns | 0.073 ns | 0.0191 | 120 B
PowerOfTen | .NET 9.0 | .NET 9.0 | 54 | 41.84 ns | 0.196 ns | 0.183 ns | 0.0052 | 88 B
BigIntegerPow | .NET 9.0 | .NET 9.0 | 54 | 118.59 ns | 0.669 ns | 0.625 ns | 0.0029 | 48 B
PowerOfTen | .NET Framework 4.8 | .NET Framework 4.8 | 54 | 98.48 ns | 0.449 ns | 0.375 ns | 0.0139 | 88 B
BigIntegerPow | .NET Framework 4.8 | .NET Framework 4.8 | 54 | 133.93 ns | 0.686 ns | 0.641 ns | 0.0317 | 201 B


</body>
</html>

lipchev:
>     * updating the benchmark results (including the `[MethodImpl(MethodImplOptions.AggressiveInlining)]` which was previously missing)
This does appear to have an effect, and since I didn't have it the last time I ran the benchmarks, I was able to observe a difference (present even after increasing the body size):
image

And here are the % difference without reduction:
image

>     * increasing the number of powers tested in the the Rounding benchmarks
I didn't bother checking the differences, but here is what the absolute values look like when reduced:

image

And here are the results without reducing:
image

src/Fractions/Fraction.Math.cs
danm-de: I don't currently see any use case for this function (outside of Fraction itself). If it's needed, we can make the method public.

PR #111, opened 2025-03-17
TryParse: replaced the usages of BigInteger.Power(TEN, exponent)

lipchev:
- `TryParse`: replaced the usages of `BigInteger.Power(TEN, exponent)` with the `PowerOfTen` function

### DIFF
--- a/src/Fractions/Fraction.TryParse.cs
+++ b/src/Fractions/Fraction.TryParse.cs
@@ -711,7 +711,7 @@ public readonly partial struct Fraction {
 
         // 4. construct the fractional part using the corresponding decimal power for the denominator
         var nbDecimals = fractionalSpan.Length;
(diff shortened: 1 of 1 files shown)


lipchev:
I started this PR yesterday, thinking about completing the coverage (I've got  everything covered at 100% in my other project)- but didn't have the time to do it.
Also, I left the benchmark running last night, just to find it today finished with `NET9` as the target (the benchmark annotation was there from before), so I wont post them here - there aren't any test values that would show a difference anyway..

@danm-de  Given all the PRs I've posted so far, there is just one more `BigInteger.Pow(TEN)` left, in the `ToDecimal` function, which is already being touched in #96. I'll probably just add it there, but if you want to make a release without it, please make sure to change it as well.

